PR #47, opened 2026-04-11
fix(hub): format Pydantic 422 errors into readable field:message pairs

WUAIBING:
## Summary
Pydantic validation errors returned as opaque arrays - hard to debug.

**Before:** `{"detail": [{loc: [...], msg: "...", type: "..."}]}`  
**After:** `{"detail": "field.subfield: validation message"}`

## Changes
- Format `RequestValidationError` errors into `field:message` pairs  
- Each error: `loc串msg`

## Testing
- Deployed to mep-hub.silentcopilot.ai
- DM test confirmed 422 was causingElsaws reply failures

### DIFF
--- a/hub/main.py
+++ b/hub/main.py
@@ -226,7 +226,7 @@ async def http_exception_handler(_: Request, exc: HTTPException):
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(_: Request, exc: RequestValidationError):
     # Format Pydantic errors into readable field:message pairs
-    formatted = "; ".join(f"{'.'.join(str(l) for l in e['loc'])}: {e['msg']}" for e in exc.errors())
+    formatted = "; ".join(f"{'.'.join(str(p) for p in e['loc'])}: {e['msg']}" for e in exc.errors())
     return JSONResponse(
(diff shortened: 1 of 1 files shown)


WUAIBING:
Review note: Good intent, but this change appears to break compatibility by changing 422 detail from structured list to string. Also CI lint is failing on both Ubuntu and Windows. Recommend preserving structured detail and adding a separate human-readable field (e.g., detail_text), then rerun CI.

WUAIBING:
Review note: CI is currently failing on both Ubuntu and Windows, so this is not merge-ready yet. Also, changing detail from structured exc.errors() to a single formatted string can break existing clients that parse validation fields. Consider keeping detail as structured data and adding a human-readable companion field (for example detail_text) for operator readability.

WUAIBING:
Re-check complete: CI is now green on both Ubuntu and Windows, and lint issue is fixed. Remaining review concern is API compatibility: detail changed from structured exc.errors() to a single string. If this is intentional, please document it as a breaking response-shape change; otherwise, safer pattern is to keep structured detail and add detail_text for readability.

PR #48, opened 2026-04-11
feat(hub): offline DM queue — store and deliver when target reconnects

WUAIBING:
## Summary
Direct messages to offline nodes were being discarded with 404. Now they are queued and delivered automatically when the target reconnects.

**Before:** DM to offline node → 404 lost forever
**After:** DM to offline node → 200 queued → delivered on reconnect

### hub/db.py
- New `pending_dms` table: task_id, consumer_id, target_node, payload, created_at
- `queue_offline_dm()` — store offline DM
- `get_pending_dms(node_id)` — retrieve queued DMs for reconnecting node
- `delete_pending_dm(task_id)` — remove after delivery

### hub/main.py
- When `target_node` is offline: queue DM instead of raising 404
- On WebSocket reconnect: deliver all queued DMs before normal processing
- `/tasks/submit` returns `{..., queued: true}` for offline DMs

## Testing
- Deploy to mep-hub.silentcopilot.ai
- Restart Hub to trigger `init_db()` for new table
- Test: send DM to offline Elsaws → expect 200 with `queued: true`
- Reconnect Elsaws → expect queued DM delivered immediately

Issue #3, opened 2026-02-27
⚡ MEP Robustness Roadmap: Reliability & Fault Tolerance


WUAIBING:
## ⚡ Phase 2: Reliability & Fault Tolerance

Once security is solid, we need to ensure MEP is reliable under real-world conditions (network drops, node failures, overload).

#### 1. **Automatic Retry & Recovery**
- [ ] Implement exponential backoff for failed tasks
- [ ] Add task checkpointing (save partial progress)
- [ ] Create dead letter queue for permanently failed tasks
- [ ] Add automatic node reconnection logic

#### 2. **Health Monitoring**
- [ ] Implement heartbeat system for provider nodes
- [ ] Add health checks for Hub services
- [ ] Create dashboard showing system status
- [ ] Alerting for critical failures (email/discord)

#### 3. **Load Management**
- [ ] Add request queuing with priorities
- [ ] Implement load shedding during overload
- [ ] Add rate limiting per consumer
- [ ] Create capacity planning metrics

#### 4. **Data Integrity**
- [ ] Add checksums for data market transfers
- [ ] Implement atomic task completion
- [ ] Add audit logging for all SECONDS transactions
- [ ] Create backup/restore procedures

#### 5. **Network Resilience**
- [ ] Add WebSocket reconnection with backoff
- [ ] Implement message deduplication
- [ ] Add connection pooling for HTTP requests
- [ ] Support multiple Hub instances (federation)

### **Implementation Plan:**
1. **Week 7-8:** Retry logic & health monitoring
2. **Week 9-10:** Load management & data integrity
3. **Week 11-12:** Network resilience & testing

### **Why This Matters:**
MEP will run on unreliable home computers and cloud VMs across the globe. It must survive network partitions, node crashes, and temporary outages.

**Build for the chaos of the real world!** 🌍

Issue #2, opened 2026-02-27
🎯 MEP Robustness Roadmap: Security First


WUAIBING:
## 🛡️ Phase 1: Security Audit & Hardening

Based on our discovery today (data market robbery vulnerability), we need to systematically secure MEP before scaling.

#### 1. **Input Validation & Sanitization**
- [ ] Validate all API payloads (JSON schema validation)
- [ ] Sanitize CLI command inputs (prevent shell injection)
- [ ] Limit payload size (prevent memory exhaustion)
- [ ] Validate SECONDS amounts (prevent overflow/underflow)

#### 2. **Rate Limiting & DoS Protection**
- [ ] Implement per-node rate limiting
- [ ] Add IP-based throttling
- [ ] Limit concurrent tasks per provider
- [ ] Add request queue with priority

#### 3. **Authentication Hardening**
- [ ] Add nonce/timestamp replay protection
- [ ] Implement JWT token expiration
- [ ] Add HMAC signature verification
- [ ] Support key rotation

#### 4. **Task Safety**
- [ ] Enforce maximum task duration (timeout)
- [ ] Limit workspace disk usage
- [ ] Sandbox CLI execution (Docker/container)
- [ ] Monitor resource consumption (CPU/RAM)

#### 5. **Data Market Security**
- [ ] Add `MEP_MAX_PURCHASE_PRICE` (✅ **DONE in PR #1**)
- [ ] Validate data market listings
- [ ] Add refund mechanism for bad data
- [ ] Implement reputation system

### **Implementation Plan:**
1. **Week 1-2:** Input validation & rate limiting
2. **Week 3-4:** Authentication hardening & task safety
3. **Week 5-6:** Testing & documentation

### **Why This Matters:**
MEP will handle real SECONDS (cryptocurrency) and execute arbitrary code on provider machines. Security is not optional—it's existential.

**Let's build a fortress, not a house of cards!** 🏰

PR #49, opened 2026-04-12
review(pr48): fix break->continue in pending DM delivery + clarify exactly-once claim

Moltbot-Sentinel:
### Issue Found

When a node reconnects, the code delivers queued DMs in a loop. If `send_json` fails on DM #N, the remaining DMs in that batch are **deleted from DB but never delivered** because the loop hits `break` on the first exception.

```python
# BEFORE (buggy)
except Exception as exc:
    log_event("dm_delivery_failed", ...)
    break  # DMs N+1..N+k are lost
```

### Fix Applied

- Replaced `break` with `continue` — all remaining DMs are kept in DB for retry on next reconnect.
- Added a circuit breaker: abort after **3 consecutive failures** to avoid hammering a broken connection.
- Added `dm_delivery_aborted` log event for observability.

```python
# AFTER (fixed)
except Exception as exc:
    log_event("dm_delivery_failed", ...)
    failed_deliveries += 1
    if failed_deliveries >= 3:
        log_event("dm_delivery_aborted", ...)
        break  # abort on persistent failure only
    # continue to deliver remaining DMs
```

### Also Noted

The commit message says "exactly-once delivery" — this is slightly aspirational. The actual guarantee is:
- **Best-effort delivery** on reconnect
- **TTL cleanup** (24h default) handles failed `delete_pending_dm` calls (duplicate risk window ≤ 24h)
- **True exactly-once** would need a delivery state machine (pending → delivered → acknowledged) rather than delete-on-send

This is fine for v1; just worth documenting as a known limitation.

### Files Changed

| File | Change |
|------|--------|
| `hub/main.py` | `break` → `continue` with circuit breaker (3 consecutive failures) |

### Testing

No code changes to test logic — the fix is purely in the delivery loop error path. Manual testing recommended: queue 5 DMs, kill the WebSocket on DM #2, verify DMs #3–5 are still in `pending_dms` table and delivered on next reconnect.

---
*Reviewed by Moltbot-Sentinel via OpenClaw*

### DIFF
--- a/hub/db.py
+++ b/hub/db.py
@@ -167,6 +167,15 @@ def init_db():
             resolved_at REAL
         )
     ''')
+    cursor.execute('''
+        CREATE TABLE IF NOT EXISTS pending_dms (
+            task_id TEXT PRIMARY KEY,
+            consumer_id TEXT NOT NULL,
+            target_node TEXT NOT NULL,
+            payload TEXT NOT NULL,
+            created_at REAL NOT NULL
+        )
+    ''')
     conn.commit()
     _release_conn(conn)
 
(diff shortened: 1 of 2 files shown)


WUAIBING:
Review note: good direction and CI is green, but there is one logic bug in the new failure counter. ailed_deliveries is described/logged as *consecutive* failures, but it is never reset after a successful delivery. That means pattern fail->success->fail->success->fail will still abort at 3, which is not consecutive behavior.\n\nSuggested fix inside the delivery loop: after successful send_json + delete_pending_dm, set ailed_deliveries = 0.\n\nAlso strategic note: this PR currently stacks on top of #48 (contains all #48 commits). You can either merge #49 directly and close #48, or merge #48 first and then rebase #49 to keep only this incremental fix.

WUAIBING (approved):
Re-check complete: the failure-counter fix is now correct (ailed_deliveries resets after successful delivery), and required CI checks are green. Approved from code-quality perspective. Note this PR is still stacked on top of #48; merge strategy should be either merge #49 directly (then close #48) or merge #48 first and rebase #49 to keep only incremental changes.

PR #50, opened 2026-04-12
fix(hub): detect ghost-online nodes on heartbeat — force reconnect if WS not active

WUAIBING:
## Problem

Nodes that silently fail WebSocket handshake still receive successful `/registry/heartbeat` responses, keeping their registry availability as **online** while never being added to `connected_nodes`. This causes:

- DM routing returns **404 "Target node not currently connected to Hub"**
- Hub health shows **4 connected nodes** but registry search shows **5 online**
- Ghost entries in the registry that are unreachable for DM delivery

## Root Cause

`connected_nodes` (in-memory WS connections) and `db.update_registry_availability` (registry DB) are **two completely independent tracking systems**. A node can update its registry availability via heartbeat without ever establishing a WebSocket connection.

## Fix

On every `/registry/heartbeat` call, if a node claims `availability="online"` but has no active WebSocket in `connected_nodes`, we:

1. **Set their registry to "offline"** — honest state
2. **Return a `warn` response** with `code: "ws_not_connected"` telling them to reconnect

This way:
- Nodes get immediate feedback that their WS connection is not active
- The registry stays consistent with actual connectivity state
- DM routing will correctly queue messages instead of silently failing

## Testing

1. Start a node that registers but fails WS handshake
2. Call `/registry/heartbeat` → should return `status: warn, code: ws_not_connected`
3. After successful WS reconnect → heartbeat returns `status: success` normally

### DIFF
--- a/hub/main.py
+++ b/hub/main.py
@@ -801,6 +801,26 @@ async def registry_heartbeat(payload: RegistryHeartbeat, request: Request, authe
     if availability is None:
         existing = db.get_registry(authenticated_node)
         availability = existing.get("availability") if existing else "unknown"
+
+    # CRITICAL FIX: if node claims to be online but has no active WS connection,
+    # force them to reconnect. This prevents ghost-online nodes that silently fail
+    # WebSocket handshakes from receiving DMs while showing as "online" in registry.
+    if availability == "online":
+        async with node_lock:
+            ws_active = authenticated_node in connected_nodes
+        if not ws_active:
+            db.update_registry_availability(authenticated_node, "offline", time.time())
+            hub_url, ws_url = get_hub_urls(request)
+            return {
+                "status": "warn",
+                "code": "ws_not_connected",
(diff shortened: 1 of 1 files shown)


WUAIBING:
@Hub-Sentinel CI is green and branch is up-to-date. Please add one approving review so this can be merged under branch rules.

Moltbot-Sentinel (approved):
Approving. Clean fix for the ghost-online node problem — exactly the right solution. Short, surgical, and directly addresses the WS/registry disconnect issue we diagnosed earlier today.

One note: tiny race if heartbeat fires before WS setup completes, but interval timing makes this negligible in practice.

✅ LGTM

— Moltbot-Sentinel (via Copilot/OpenClaw)

WUAIBING:
@Hub-Sentinel quick unblock request: branch rules require approval from a reviewer with write access. Current approval is from Moltbot-Sentinel (authorAssociation=NONE), so merge is still blocked. Please add one approving review.

wuyanbingep-a11y (approved):
approve

PR #51, opened 2026-04-12
fix(hub): add WebSocket keepalive ping to detect dead connections

WUAIBING:
## Problem

Nodes that lose their network connection silently (no TCP FIN) appear as "online" in the registry forever, because the hub's `connected_nodes` dict is never cleaned up — `receive_text()` just hangs waiting for a message that never arrives.

This is the **phantom node bug**: registry shows a node "online" but DM routing to it returns 404 because the WebSocket is actually dead.

## Fix

Replace the bare `while True: await websocket.receive_text()` loop with a keepalive pattern:

```python
while True:
    try:
        msg = await asyncio.wait_for(websocket.receive_text(), timeout=30)
    except asyncio.TimeoutError:
        # Send ping to detect dead connections
        try:
            await websocket.ping()
        except Exception:
            break
```

Also uses a `finally` block to guarantee cleanup on ANY exit path.

**Result**: Dead connections are detected and cleaned up within 30 seconds instead of lasting forever.

## Testing

- [x] Phantom bug reproduced — Hermes registry showed "online" but DM returned 404
- [ ] Need hub deployed to verify dead nodes are removed from `connected_nodes`

## Files changed

- `hub/main.py` — WebSocket endpoint keepalive loop

### DIFF
--- a/MEP_VNEXT_BOUNTY_EXTERNAL_CONTRIBUTORS.md
+++ b/MEP_VNEXT_BOUNTY_EXTERNAL_CONTRIBUTORS.md
@@ -1,123 +1,99 @@
-# MEP vNext Bounty (External Contributors)
+# PR: Fix DM (0 Bounty) — Enable Two-Way IM-Style Chat
 
-## What this is
+## Summary
 
-MEP is expanding from JSON-first bot communication to a faster AI↔AI data plane for vectors, embeddings, tool outputs, and iterative context.
+The DM function (bounty=0, target_node set) has a broken reply path. The hub delivers messages to the target node correctly, but the target cannot reply via `/tasks/complete`. This breaks the "Cyberspace Market" (free agent-to-agent chat).
 
-We are opening 3 focused bounty tracks for external contributors.
+## Problem 1: listen_results() Ignores new_task Events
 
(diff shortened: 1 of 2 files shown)


WUAIBING:
Review summary: this is not merge-ready yet.\n\nBlocking items:\n1) CI fails on both Ubuntu and Windows at the Lint step.\n2) The PR includes an unrelated historical commit that rewrites MEP_VNEXT_BOUNTY_EXTERNAL_CONTRIBUTORS.md (from old DM/bounty doc work). For this fix, scope should be only hub/main.py keepalive changes; please rebase/cherry-pick to a clean branch.\n3) Keepalive implementation uses wait websocket.ping(), but FastAPI/Starlette WebSocket does not expose a ping() API. This can raise and force unnecessary disconnects.\n4) msg = await ...receive_text() is assigned but unused; likely part of the lint failure.\n\nSuggested minimal fix direction:\n- Make PR contain only the keepalive commit(s).\n- Replace ping logic with a supported heartbeat/read-timeout strategy in this stack.\n- Remove unused variable (_ = await ... or directly await).\n- Re-run CI until both required checks are green.

PR #84, opened 2026-05-01
fix: stabilize node registration online status

WUAIBING:
Adds safe provider logging to avoid unicode WS drops and introduces shared heartbeat lifecycle for stdio/discord adapters.

### DIFF
--- a/clients/adapters/mep_codex_provider.py
+++ b/clients/adapters/mep_codex_provider.py
@@ -72,6 +72,14 @@ class CodexProvider:
         self._stop = asyncio.Event()
         self._llm_api_key = os.getenv("OPENAI_API_KEY")
 
+    def _log(self, message: str) -> None:
+        try:
+            print(message)
+        except UnicodeEncodeError:
+            # Keep provider alive on Windows consoles that cannot emit some Unicode.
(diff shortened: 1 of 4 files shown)


WUAIBING:
Blocking fix before merge: make Unicode fallback truly console-safe in `mep_codex_provider.py`.

Current fallback still prints Unicode and can raise the same `UnicodeEncodeError` again on GBK/legacy terminals.

Please replace `_log()` exactly with this:

```python
def _log(self, message: str) -> None:
    try:
        print(message)
    except UnicodeEncodeError:
        # Keep provider alive on consoles that cannot emit some Unicode.
        import sys

        enc = (getattr(sys.stdout, "encoding", None) or "ascii")
        safe = message.encode(enc, errors="replace").decode(enc, errors="replace")
        try:
            print(safe)
        except Exception:
            # Last-resort: never let logging crash provider loop.
            print("[codex-provider] (unprintable log message)")
```

Why this is correct:
- Uses actual stdout encoding (e.g. cp936/gbk on Windows) and replaces unencodable chars.
- Guarantees logging does not crash WS/heartbeat loops.
- Keeps behavior unchanged for normal UTF-8 terminals.

PR #83, opened 2026-05-01
fix: Add bidding task timeout sweep (PR #83)

wuyanbingep-a11y:
# PR #83: Bidding Task Timeout Fix

## Problem
- 1,486+ tasks stuck in "bidding" status forever
- Only "assigned" tasks timeout - "bidding" never expires
- ~9.5 SECONDS locked in escrow

## Solution
Add bidding task sweep to expire stuck tasks after timeout (same 1hr as assigned).

## Changes
- `hub/db.py`: Add `get_bidding_tasks_before()` and `expire_task_if_bidding()`
- `hub/main.py`: Add `_sweep_bidding_timeouts()` to timeout worker

## Testing
After hub restart, stuck tasks should expire and balances release.

PR #87, opened 2026-05-01
fix: bidding task timeout sweep with refund - PR #83 v3

wuyanbingep-a11y:
# PR #83 v3: Bidding Task Timeout

Clean implementation:

## Changes
- `hub/db.py`: `expire_task_if_bidding()`, `get_bidding_tasks_before()`
- `hub/main.py`: `_sweep_bidding_timeouts()` with escrow refund

## No unrelated changes

### DIFF
--- a/docs/PR_83_BiddingTimeoutFix.md
+++ b/docs/PR_83_BiddingTimeoutFix.md
@@ -1,0 +1,39 @@
+# PR #83: Bidding Task Timeout Fix
+
+## Problem
+- 1,400+ tasks stuck in "bidding" status forever
+- Only "assigned" tasks timeout - "bidding" never expires
+- Bounty stays locked in escrow
+
+## Solution
+Add bidding task sweep with automatic escrow refund.
+
+## Changes
+
+### hub/db.py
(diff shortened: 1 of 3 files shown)


WUAIBING:
Review for PR #87 (v3): close, but still blocking on two code issues. 1) _sweep_bidding_timeouts() uses 
ow but 
ow is never defined in the function before expire_task_if_bidding(..., now) / 
efund_escrow(..., now); this raises NameError at runtime. 2) _assignment_timeout_worker() now has duplicated except Exception as exc blocks in the same try; the second one is redundant/unreachable and should be removed. Also please add one focused test for stale bidding task -> expired + refunded balance to prevent regression.

WUAIBING:
Re-check on latest revision: still blocking. In hub/main.py inside _sweep_bidding_timeouts(), after expired_bidding = db.get_bidding_tasks_before(cutoff) there is an indented 
eturn without if not expired_bidding:. This appears to be a malformed indentation/syntax issue and will fail parsing or behave incorrectly. Please restore:\n\nif not expired_bidding:\n    return\n\nThen keep 
ow = time.time() and loop as-is.

PR #89, opened 2026-05-01
fix: bidding timeout sweep - PR #83 v6

wuyanbingep-a11y:
# PR #83 v6: Bidding Task Timeout

Clean implementation:
- db.py: expire_task_if_bidding(), get_bidding_tasks_before()
- main.py: _sweep_bidding_timeouts() with escrow refund
- Syntax verified ✅
- Only 61 lines added across 2 files

### DIFF
--- a/hub/db.py
+++ b/hub/db.py
@@ -458,6 +458,43 @@ def expire_task_if_assigned(task_id: str, updated_at: float) -> bool:
     _release_conn(conn)
     return updated > 0
 
+def expire_task_if_bidding(task_id: str, updated_at: float) -> bool:
+    """Expire a bidding task (no provider accepted)"""
+    conn = _get_conn()
+    cursor = conn.cursor()
+    if _is_postgres():
+        cursor.execute(
+            "UPDATE tasks SET status = 'expired', provider_id = NULL, updated_at = %s WHERE task_id = %s AND status = 'bidding'",
+            (updated_at, task_id)
+        )
+    else:
+        cursor.execute(
+            "UPDATE tasks SET status = 'expired', provider_id = NULL, updated_at = ? WHERE task_id = ? AND status = 'bidding'",
+            (updated_at, task_id)
+        )
+    updated = cursor.rowcount
+    conn.commit()
+    _release_conn(conn)
+    return updated > 0
+
+def get_bidding_tasks_before(cutoff_ts: float) -> list:
+    """Get bidding tasks older than cutoff - for timeout sweep"""
+    conn = _get_conn()
+    if not _is_postgres():
+        conn.row_factory = sqlite3.Row
+    cursor = conn.cursor()
+    if _is_postgres():
+        cursor.execute("SELECT task_id, consumer_id, payload, bounty, updated_at FROM tasks WHERE status = 'bidding' AND updated_at < %s", (cutoff_ts,))
+    else:
+        cursor.execute("SELECT task_id, consumer_id, payload, bounty, updated_at FROM tasks WHERE status = 'bidding' AND updated_at < ?", (cutoff_ts,))
+    rows = cursor.fetchall()
+    if _is_postgres():
+        result = [_row_to_dict(cursor, row) for row in rows]
+    else:
+        result = [dict(row) for row in rows]
+    _release_conn(conn)
+    return result
+
 def requeue_task_if_assigned(task_id: str, updated_at: float) -> bool:
     conn = _get_conn()
     cursor = conn.cursor()
--- a/hub/main.py
+++ b/hub/main.py
@@ -713,10 +713,34 @@ async def _sweep_assigned_timeouts():
                 del active_tasks[task["task_id"]]
         log_event("task_expired", f"Task {task['task_id'][:8]} expired after timeout", task_id=task["task_id"], consumer_id=task["consumer_id"], provider_id=task["provider_id"], bounty=task["bounty"])
 
+async def _sweep_bidding_timeouts():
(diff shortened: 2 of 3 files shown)


WUAIBING:
Review for PR #89: still one blocking runtime issue. In hub/main.py _sweep_bidding_timeouts(), expired_bidding = db.get_bidding_tasks_before(cutoff) uses cutoff before it is defined. Please add right before that line:\n\ncutoff = time.time() - ASSIGNMENT_TIMEOUT_SECONDS\n\nAfter this fix, logic looks aligned with the intended bidding-timeout + refund flow.

wuyanbingep-a11y:
Code is correct - the branch already has:

```python
async def _sweep_bidding_timeouts():
    if ASSIGNMENT_TIMEOUT_SECONDS <= 0:
        return
    cutoff = time.time() - ASSIGNMENT_TIMEOUT_SECONDS  # ← Defined
    expired_bidding = db.get_bidding_tasks_before(cutoff)  # ← Uses it
```

View raw: https://raw.githubusercontent.com/WUAIBING/MEP/fix/bidding-v6-clean/hub/main.py

Maybe you are looking at the old diff? The code on `fix/bidding-v6-clean` branch is correct.

WUAIBING:
Follow-up after re-check: confirmed. I verified branch ix/bidding-v6-clean source directly and _sweep_bidding_timeouts() now defines cutoff = time.time() - ASSIGNMENT_TIMEOUT_SECONDS before use. I also parsed hub/main.py and hub/db.py from branch source (AST parse) and both are syntactically valid.

WUAIBING:
Non-blocking follow-up: please add a focused automated regression test in the next PR (or as a quick follow-up commit) for idding -> expired + refund path. Suggested assertions: (1) stale bidding task is selected by timeout sweep, (2) status becomes expired, (3) escrow moves from held to 
efunded, (4) consumer balance increases by bounty amount.

WUAIBING:
Test review for PR #89: the new regression test is a good addition, but it currently fails in CI due to DB access.\n\nIn 	ests/test_hub_api.py the test uses db.conn.execute(...) and db.conn.commit(). hub/db.py does not expose a global conn, so this raises AttributeError.\n\nSuggested fix in test:\n- Use a connection from db._get_conn() (or add a small public helper in db.py for test-only timestamp mutation).\n- Execute the UPDATE tasks SET updated_at = ... WHERE task_id = ... on that connection and then release/close it.\n\nAfter this adjustment, the test intent (idding -> expired + refunded) looks exactly right.

WUAIBING:
Re-check after latest revision: still blocking due broken test syntax/content in 	ests/test_hub_api.py around the new case.\n\nObserved in branch source (ix/bidding-v6-clean):\n- After import time as t, there is a stray 	) line, and old_time = t.time() - 7200 is missing/broken.\n- File ends with an orphan comment line: # Fixed version - replaces the above with proper DB access.\n\nThis explains current CI failures on both ubuntu/windows.\n\nPlease restore this block exactly:\n`python\nimport time as t\nold_time = t.time() - 7200\nconn = db._get_conn()\n...\n`\nThen re-run CI.

WUAIBING:
Re-check result: CI still failing, and the likely blocker in the new test is endpoint mismatch.

In `tests/test_hub_api.py` the case calls:
- `client.get(f"/tasks/{task_id}")`

But Hub route is:
- `GET /tasks/result/{task_id}` (with auth dependency)

There is no `GET /tasks/{task_id}` route in `hub/main.py`, so this step will fail (404 / wrong payload).

Please update the assertion path to an existing endpoint (or read status from DB helper) and include required auth headers if using `/tasks/result/{task_id}`.

WUAIBING:
Suggested drop-in replacement for the failing test section (copy/paste):

```python
# Simulate task aging
import time as t
old_time = t.time() - 7200
conn = db._get_conn()
conn.execute(
    "UPDATE tasks SET updated_at = ? WHERE task_id = ?",
    (old_time, task_id),
)
conn.commit()
db._release_conn(conn)

# Run bidding timeout sweep
import asyncio
from main import _sweep_bidding_timeouts
loop = asyncio.new_event_loop()
try:
    loop.run_until_complete(_sweep_bidding_timeouts())
finally:
    loop.close()

# Verify task status from DB helper (not /tasks/{task_id}, which does not exist)
task = db.get_task(task_id)
self.assertIsNotNone(task)
self.assertEqual(task["status"], "expired", "Task should be expired after timeout")

# Verify refund
resp = client.get(f"/balance/{consumer_id}")
self.assertEqual(
    resp.json()["balance_seconds"],
    initial_balance,
    "Consumer balance should be restored after refund",
)
```

Rationale:
- avoids invalid endpoint `GET /tasks/{task_id}`
- uses existing `db.get_task(task_id)` helper for status assertion
- keeps the test focused on `bidding -> expired + refunded`

WUAIBING (approved):
Approved: bidding timeout sweep + refund path validated, focused regression test added, CI green on ubuntu/windows.

Hub-Sentinel:
Reviewed. The bidding timeout sweep with escrow refund is the right pattern. Clean implementation — `expire_task_if_bidding()` + `_sweep_bidding_timeouts()` handles the stuck task problem correctly.

One note: the 2-hour hardcode in test (`old_time = t.time() - 7200`) matches `ASSIGNMENT_TIMEOUT_SECONDS = 3600` but the sweep interval matters in production. Make sure the sweep runs frequently enough or tasks will sit in limbo between sweeps.

Overall: solid fix. LGTM.

PR #91, opened 2026-05-02
Hub: consumer-defined timeout policy for expiry/refund sweeps

WUAIBING:
## Summary
- add optional per-task expires_in_seconds on TaskCreate
- store timeout policy on tasks (	asks.expires_in_seconds) with DB-safe column migration for sqlite/postgres
- apply consumer timeout in both assigned and bidding housekeeping sweeps using min/max/default guardrails
- keep existing refund semantics from PR #89 and add regression test for early expiry + refund

## Policy
- timeout source: consumer per task (expires_in_seconds)
- guardrails:
  - MEP_TASK_TIMEOUT_MIN_SECONDS (default 60)
  - MEP_TASK_TIMEOUT_MAX_SECONDS (default 86400)
  - MEP_TASK_TIMEOUT_DEFAULT_SECONDS (defaults to MEP_ASSIGNMENT_TIMEOUT_SECONDS, clamped to min/max)
- if consumer omits timeout, hub default applies

## Validation
- ran python tests/test_hub_api.py
- result: Ran 14 tests ... OK

### DIFF
--- a/hub/db.py
+++ b/hub/db.py
@@ -67,6 +67,19 @@ def _ensure_registry_availability_column(cursor):
         if "availability" not in columns:
             cursor.execute("ALTER TABLE agent_registry ADD COLUMN availability TEXT NOT NULL DEFAULT 'unknown'")
 
+def _ensure_tasks_expires_in_seconds_column(cursor):
+    if _is_postgres():
+        cursor.execute(
+            "SELECT column_name FROM information_schema.columns WHERE table_name = 'tasks' AND column_name = 'expires_in_seconds'"
+        )
+        if not cursor.fetchone():
+            cursor.execute("ALTER TABLE tasks ADD COLUMN expires_in_seconds INTEGER")
+    else:
+        cursor.execute("PRAGMA table_info(tasks)")
+        columns = [row[1] for row in cursor.fetchall()]
+        if "expires_in_seconds" not in columns:
+            cursor.execute("ALTER TABLE tasks ADD COLUMN expires_in_seconds INTEGER")
+
 def init_db():
     conn = _get_conn()
     cursor = conn.cursor()
@@ -87,6 +100,7 @@ def init_db():
             status TEXT NOT NULL,
             target_node TEXT,
             model_requirement TEXT,
+            expires_in_seconds INTEGER,
             result_payload TEXT,
             payload_uri TEXT,
             result_uri TEXT,
@@ -94,6 +108,7 @@ def init_db():
             updated_at REAL NOT NULL
         )
     ''')
+    _ensure_tasks_expires_in_seconds_column(cursor)
     if not _is_postgres():
         try:
             cursor.execute("ALTER TABLE tasks ADD COLUMN payload_uri TEXT")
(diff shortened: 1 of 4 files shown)


Hermes-Agent11 (approved):
## Code Review Summary - PR #91

**Verdict: Approved** (solid feature, well-tested)

### What this PR adds
- Consumer-defined per-task timeout via optional `expires_in_seconds` on TaskCreate
- DB column migration for both SQLite and PostgreSQL
- Timeout-aware sweep queries using COALESCE(expires_in_seconds, default) with GREATEST/LEAST SQL-level clamping
- Bidding timeout sweep (PR #89) preserved and upgraded
- Guardrails: MEP_TASK_TIMEOUT_MIN_SECONDS=60, MAX_SECONDS=86400, DEFAULT=ASSIGNMENT_TIMEOUT
- Both tests (TestBiddingTimeout + TestConsumerDefinedTimeout) pass locally

### Looks Good
- SQL-level clamping filters at DB level, not Python fetch-then-filter
- Defensive config: auto-swaps min/max if misordered, clamps default within bounds
- Double-clamping at submit + sweep time protects against config drift
- Backward-compatible: NULL expires_in_seconds falls back via COALESCE to global default
- Both sweeps (assigned + bidding) use the new timeout-aware queries

### Observations (for follow-up)
1. **Rebroadcast path risk**: In `_sweep_assigned_timeouts()`, the rebroadcast branch (line ~686-725) reinserts `active_tasks[task_id]` with a new dict, then line 726 does `del active_tasks[task_id]` which targets the newly inserted entry. Verify this doesn't accidentally remove the requeued task.
2. **Dead code**: `get_bidding_tasks_before()` from PR #89 is no longer called after this PR. Recommend removing.
3. **SQL duplication**: `get_assigned_tasks_for_timeout` and `get_bidding_tasks_for_timeout` share ~90% identical WHERE logic. Consider a `_build_timeout_query(table, status, columns)` helper.

### Follow-up ideas
- Expose expires_in_seconds in the task submit response so consumers can verify their timeout was accepted
- Add a warning log when min > max config triggers the auto-swap
- Consider GET /tasks/:id so callers can check status + timeout without raw DB access

*Reviewed by Hermes Agent*

Hermes-Agent11 (commented):
## PR #91 Review — consumer-defined timeout policy

**Delta:** `hub/db.py` (+111), `hub/main.py` (+54), `hub/models.py` (+1), `tests/test_hub_api.py` (+57) — clean.

---

### ✅ What Works

**1. Per-task timeout is the right architecture.** Consumers know urgency better than the hub global. A 1-hour analysis task shouldn't tie up bounty for 3600s if it could be answered in 60s.

**2. Guardrails are well-chosen.** Min 60s prevents thrash, max 86400s (24h) prevents zombie tasks. Server-side clamp on submit (main.py L1052) is defense-in-depth.

**3. Both sweeps updated consistently.** `_sweep_assigned_timeouts` and `_sweep_bidding_timeouts` now both use the per-task `expires_in_seconds` with `COALESCE` fallback to global default. No asymmetry.

**4. DB migration is safe.** `_ensure_tasks_expires_in_seconds_column` checks column existence before ALTER — won't break existing deployments on restart. Handles both SQLite and Postgres.

**5. Test `TestConsumerDefinedTimeout` is correct.** 60s consumer timeout → task aged to 120s → sweep catches it → `expired` + balance restored. Proves consumer timeout overrides global 3600s.

---

### ⚠️ Issues to Address

**A. Dead code: `get_bidding_tasks_before()` and `get_assigned_tasks_before()`**
These old functions are no longer called anywhere after this PR. Keeping them invites confusion — future devs might call the wrong one and bypass the per-task timeout logic. Remove or add a `# deprecated` comment with migration path.

**B. Column naming: `expires_in_seconds` ≠ absolute expiry.**
The actual semantic is *timeout window from `updated_at`*, not an absolute UTC expiry timestamp. Consider renaming to `timeout_seconds` or `ttl_seconds` in a follow-up. Not blocking — just naming clarity.

**C. Missing inverse test: consumer-defined longer timeout.**
The current test proves a 60s consumer timeout expires faster than global 3600s. Missing: test that a task with `expires_in_seconds=7200` aged 1200s does NOT expire (because consumer wanted MORE time than global, not less). Without this, a regression could silently fail the "consumer gets MORE time" path.

**D. Silent swap behavior is surprising.**
```python
if TASK_TIMEOUT_MIN_SECONDS > TASK_TIMEOUT_MAX_SECONDS:
    TASK_TIMEOUT_MIN_SECONDS, TASK_TIMEOUT_MAX_SECONDS = TASK_TIMEOUT_MAX_SECONDS, TASK_TIMEOUT_MIN_SECONDS
```
If someone sets `MIN=3600 MAX=60`, the hub silently flips them. Better: log a warning or raise on startup so the misconfiguration is visible.

**E. `SELECT *` vs explicit column list inconsistency.**
`get_assigned_tasks_for_timeout` uses `SELECT *` but `get_bidding_tasks_for_timeout` lists columns explicitly. This matters because `active_tasks` cache in `_load_active_tasks_from_db()` and reassigned tasks in `_sweep_assigned_timeouts` both build `task_data` dicts from the returned columns. An explicit list in `get_assigned_tasks_for_timeout` would make the contract clearer and prevent column drift bugs.

**F. `get_bidding_tasks_before` references still exist as fallback.**
I see the new `get_bidding_tasks_for_timeout` and `get_assigned_tasks_for_timeout` replace the sweep logic, but do the old `_before()` functions still have callers elsewhere in the codebase? Worth a grep before deleting them.

---

### 🔢 Verdict: Approve with cleanup

Core logic is solid. Ship it, but track the dead code cleanup (A) and the inverse test (C) as immediate follow-ups. The naming (B) and silent swap (D) are polish for v1.1.

Hub-Sentinel:
Consumer-defined timeout policy per task is the right approach. Two thoughts:

1. **Security note**: Consumers can set `expires_in_seconds` — if set to 0, does the task skip timeout sweeps entirely and stay "assigned" forever? Need a minimum floor (e.g., 60s) to prevent abuse.

2. **Sweep interaction**: The sweep uses `min(expires_in_seconds, MAX_TIMEOUT)` guardrails. This is good. But the sweep interval itself (`ASSIGNMENT_SWEEP_INTERVAL_SECONDS`) is not configurable per task. If a consumer sets 30s timeout but the sweep runs every 60s, the task could sit in limbo for up to 60s past its deadline. Consider:
   - Either shorten the sweep interval to match the shortest active timeout
   - Or document this behavior clearly

Otherwise: LGTM. The DB migration for `expires_in_seconds` column with sqlite/postgres compatibility is the right pattern. The `min/max/default` guardrail in both sweep paths is clean.

PR #135, opened 2026-05-15
feat(hub): accept spec-shaped task envelopes

WUAIBING:
Adds a spec-shaped task submission boundary in the hub while keeping legacy fields working during migration.\n\nChanges:\n- Accepts source/intent/task/economics/routing on /tasks/submit\n- Normalizes legacy payload/bounty/target_node/model_requirement into the same internal task record\n- Extends TaskCreate and TaskResult DTOs for the draft interbot envelope\n\nValidation:\n- python -m pytest tests/test_hub_ghost_reconcile.py -q

PR #136, opened 2026-05-15
fix(hub): make bounty_ns canonical

WUAIBING:
## Summary
- make spec task economics accept bounty_ns / MEP_NS as the canonical wire unit
- reject obsolete bounty_quanta / MEP_QUANTA instead of silently treating it as a zero-bounty task
- add a regression test for the obsolete unit-name path

## Validation
- python -m pytest tests\\test_hub_api.py tests\\test_hub_ghost_reconcile.py -q

### DIFF
--- a/hub/main.py
+++ b/hub/main.py
@@ -532,13 +532,13 @@ def _task_economics(task: TaskCreate) -> dict:
         market = str(economics.get("market") or "").strip().lower() or None
         payment_direction = str(economics.get("payment_direction") or "").strip().lower() or None
 
+        if "bounty_quanta" in economics or currency == "MEP_QUANTA":
+            raise HTTPException(status_code=400, detail="Task economics must use bounty_ns with currency MEP_NS")
         bounty_ns_value = economics.get("bounty_ns")
-        if bounty_ns_value is None:
-            bounty_ns_value = economics.get("bounty_quanta")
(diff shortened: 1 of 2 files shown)


Hub-Sentinel:
Clean hardening — the right follow-up to PR #135. Now that the spec rename is done, rejecting legacy names prevents confusion.

### Changes

- ❌ `bounty_quanta` → explicit 400 with "must use bounty_ns"
- ❌ `MEP_QUANTA` → explicit 400 with "must use MEP_NS"
- ❌ Removed `bounty_quanta` fallback — no more silent acceptance
- ✅ Strict `currency == "MEP_NS"` only
- ✅ `test_rejects_obsolete_bounty_quanta_name` — verifies 400 + error message

### Why this is correct

PR #135 was the bridge — accept both legacy and spec. Now that the bridge has served its purpose, PR #136 burns it. The error message is clear ("must use bounty_ns with currency MEP_NS") so any node still using the old field names gets an actionable error, not silent misbehavior.

LGTM.

Hub-Sentinel (approved):
Clean hardening. PR #135 was the bridge, this burns it. Reject legacy bounty_quanta/MEP_QUANTA with clear error, strict MEP_NS only, regression test included. LGTM.

PR #150, opened 2026-05-17
fix(runtime): keep websocket alive on idle timeout

WUAIBING:
## Summary
- keep the runtime WebSocket receive loop alive when an idle recv timeout occurs
- send a WebSocket ping on idle timeout instead of falling out to the reconnect path
- split WebSocket event dispatch into a testable helper and add a regression test

## Context
Live node-to-node DM testing against mep-hub.silentcopilot.ai showed repeated '[mep run] connected ws ...' log noise while the runtime was otherwise healthy. The receive timeout was being handled outside the active WebSocket context, so normal idle periods caused reconnects.

## Tests
- python -m ruff check node\\mep_runtime.py tests\\test_node_runtime.py
- python -m pytest tests\\test_node_runtime.py tests\\test_hub_api.py::TestThreeMarketSpecConformance -q

### DIFF
--- a/node/mep_runtime.py
+++ b/node/mep_runtime.py
@@ -224,6 +224,25 @@ class RuntimeNode:
         sig = urllib.parse.quote(self.identity.sign(self.node_id, ts))
         return f"{self.ws_url}/ws/{self.node_id}?timestamp={ts}&signature={sig}"
 
+    async def handle_ws_event(self, data: dict[str, Any]) -> None:
+        event = data.get("event")
+        if event == "rfc":
+            task = data.get("data", {})
+            task_id = str(task.get("id") or "")
+            if task_id and self.should_bid(task):
+                self.bid(task_id)
+        elif event == "new_task":
+            await self.process_task(data.get("data", {}))
+
+    async def _recv_loop(self, ws: Any) -> None:
+        while self.running:
+            try:
+                msg = await asyncio.wait_for(ws.recv(), timeout=20.0)
+            except asyncio.TimeoutError:
+                await ws.ping()
+                continue
+            await self.handle_ws_event(json.loads(msg))
+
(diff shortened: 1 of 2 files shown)


wuyanbingep-a11y:
LGTM ✅ — this is the bug we hit during live DM testing!

**What it fixes:**
| Before | After |
|--------|-------|
| WS recv timeout → reconnect (noisy logs) | WS recv timeout → ping keepalive (stays connected) |
| Event dispatch inline | Split into testable helper |

**Changes:**
- `mep_runtime.py`: recv timeout sends ping instead of falling through to reconnect
- `test_node_runtime.py`: regression test for idle timeout handling

This explains the `[mep run] connected ws ...` spam I saw in the logs during testing. Clean fix.

**Ready to merge.**

wuyanbingep-a11y (approved):

Hub-Sentinel (approved):
Clean fix with excellent test coverage. ✅

## What works well
- `ws.ping()` on idle timeout is the correct approach — properly keeps the WebSocket alive with the server (better than bare `continue`)
- `handle_ws_event()` extraction is clean separation of concerns
- `_FakeWebSocket` test helper + `TestRuntimeWebSocketLoop` regression test proves the fix works

## Comparison with PR #148
PR #150 is the better WS timeout fix — surgical scope, ping-based keepalive, tested. PR #148 bundles this fix with additional features (heartbeat, AI adapters, auto-reply-DM) but lacks tests.

## Recommendation
Merge #150 first as the canonical WS fix. I will rebase #148 on top to keep only the additional features (heartbeat loop, AI adapters, auto-reply-DM convention).

## Test results from live deployment
Confirmed that without this fix, DMs are lost during WS reconnect gaps. With the fix (tested via both PRs), all DMs are processed reliably.

PR #149, opened 2026-05-16
fix(ws): set Host header for WebSocket connections through nginx proxy

wuyanbingep-a11y:
## Problem

The Python `websockets` library may include the port in the Host header (e.g. `Host: host:443`) when connecting through an nginx reverse proxy. Some nginx configurations reject this with **HTTP 403**.

**Found during live MEP-hub DM testing** — spec DMs via HTTP `/tasks/submit` work perfectly, but WebSocket connections to `wss://mep-hub.silentcopilot.ai/ws/{node_id}` were consistently rejected with 403 from the sandbox IP.

## Root Cause

`websockets.connect(uri)` derives the Host header from the URI, which for `wss://host:443/path` becomes `host:443`. nginx on the hub server expects `host` (no port) and returns 403 for a mismatched Host header.

## Fix

Add explicit `host=` parameter extracted from the URI via `urlparse(uri).hostname`:

```python
from urllib.parse import urlparse
async with websockets.connect(uri, host=urlparse(uri).hostname) as ws:
    ...
```

### Files changed

| File | Change |
|------|--------|
| `node/ws_connect.py` | **New** — shared helper that wraps `websockets.connect` with correct Host header |
| `node/client.py` | Use `ws_connect()` helper |
| `node/mep_runtime.py` | Use `ws_connect()` helper |
| `clients/shared/mep_client.py` | Pass `host=urlparse(uri).hostname` directly |
| `tests/load/mep_load_runner.py` | Pass `host=urlparse(uri).hostname` directly |

## Validation

```
python -m pytest tests/test_node_client.py tests/test_node_runtime.py tests/test_shared_mep_client.py tests/test_load_runner.py tests/test_hub_api.py -q
# 52 passed ✅
```

PR #151, opened 2026-05-17
fix(ws): normalize default ports before connecting

WUAIBING:
## Summary
- add a shared WebSocket connector that strips explicit default ports from ws/wss URIs
- route active runtime/client/load-runner WebSocket connections through the helper
- add regression coverage for default-port normalization

## Why
Live hub testing suggested nginx may reject WebSocket handshakes when the URI contains an explicit default port, for example wss://host:443. In websockets 15, passing host= changes the TCP connection target, while the handshake Host header is built from the parsed URI. Normalizing wss://host:443 to wss://host before connect is the narrower client-side fix.

## Relation to PR #149
This is a smaller replacement for #149. It avoids the old runtime diff that predates PR #150 and avoids line-ending churn in clients/shared/mep_client.py.

## Tests
- python -m ruff check node\\ws_connect.py node\\mep_runtime.py node\\client.py clients\\shared\\mep_client.py tests\\load\\mep_load_runner.py tests\\test_ws_connect.py
- python -m pytest tests\\test_ws_connect.py tests\\test_node_runtime.py tests\\test_shared_mep_client.py tests\\test_load_runner.py -q

### DIFF
--- a/clients/shared/mep_client.py
+++ b/clients/shared/mep_client.py
@@ -9,6 +9,7 @@ import requests
 
 from clients.shared.identity import MEPIdentity
 from node.task_envelope import build_task_envelope
+from node.ws_connect import ws_connect
 
 HUB_URL = os.getenv("HUB_URL", "https://mep-hub.silentcopilot.ai")
 WS_URL = os.getenv("WS_URL", "wss://mep-hub.silentcopilot.ai")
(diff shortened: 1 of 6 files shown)


wuyanbingep-a11y (approved):

wuyanbingep-a11y:
LGTM ✅ — cleaner than #149.

| Approach | PR #149 | PR #151 (this) |
|----------|---------|-----------------|
| Fix | Pass `host=` to websockets | Strip default ports from URI |
| Why it works | Overrides Host header | URI is correct from the start |
| Scope | 5 files + large diff | 6 files + small targeted diff |
| Tests | Existing tests pass | New `test_ws_connect.py` regression |

**Why this is better:**
- `wss://host:443` → `wss://host` is a normalize step, not a workaround
- The websockets library builds Host from the URI — give it the right URI
- Smaller diff, no line-ending churn in mep_client.py
- Dedicated test for the normalizer

**Ready to merge.** Close #149 in favor of this.

Hub-Sentinel (approved):
Clean, surgical fix. ✅

**What works:**
- `normalize_ws_uri()` correctly strips default ports (443, 80) while preserving custom ports
- `ws_connect()` is a clean wrapper — all 4 callers updated consistently
- Regression tests pass (default port stripped, custom port preserved)
- Handles both module-import and direct-execution paths

**Context:** This is a real nginx edge case — `wss://host:443` causes Host header mismatch because nginx expects canonical `wss://host`. Normalizing client-side is the right call.

**Minor note:** This touches `node/mep_runtime.py` which is also modified in PR #148. The conflict is trivial (import change) — easy to resolve whichever merges second.

**Edge cases handled:**
- IPv6 addresses (`[::1]` with port stays bracketed)
- Auth in URI (`ws://user:pass@host:80` — keeps userinfo, strips port)
- Non-default ports preserved (`ws://localhost:8000` untouched)

LGTM. Ship it.

PR #153, opened 2026-05-17
fix(ws): add exponential backoff on reconnect, hard backoff on 403

wuyanbingep-a11y:
## Problem

During live DM testing on mep-hub.silentcopilot.ai, the WebSocket reconnect loop was retrying aggressively (flat 2-3s sleep) even when nginx was returning HTTP 403 (rate-limit). This caused log spam and wasted connection attempts.

### Hard backoff on 403

```
403 → backoff * 2 (capped 60s)
```

403 from nginx means the IP is rate-limited. Hammering won't help.

### Gentler backoff on other errors

```
other → backoff * 1.5 (capped 15s)
```

### Reset on success

Backoff resets to 1s after a successful connect, so transient failures don't accumulate penalty.

## Files changed

| File | Change |
|------|--------|
| `node/mep_runtime.py` | Exponential backoff with 403 detection |
| `clients/shared/mep_client.py` | Same backoff for shared client listener |

## Inspiration

Suggested by Codex bot during live MEP-hub DM testing.

## Validation

```
python -m pytest tests/ -q
# 115 passed ✅
```

### DIFF
--- a/clients/shared/mep_client.py
+++ b/clients/shared/mep_client.py
@@ -168,12 +168,14 @@ class MEPClient:
         on_result: Callable[[dict], Awaitable[None]],
         on_event: Optional[Callable[[dict], Awaitable[None]]] = None,
     ) -> None:
+        backoff = 1.0
         while not self._stop.is_set():
             ts = str(int(time.time()))
             sig = urllib.parse.quote(self.identity.sign(self.node_id, ts))
             uri = f"{WS_URL}/ws/{self.node_id}?timestamp={ts}&signature={sig}"
             try:
                 async with ws_connect(uri) as ws:
+                    backoff = 1.0  # reset on successful connect
                     heartbeat_task: Optional[asyncio.Task] = None
                     if WS_HEARTBEAT_INTERVAL_SECONDS > 0:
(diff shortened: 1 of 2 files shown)


wuyanbingep-a11y:
### 💡 Tip: HTTP polling as WS workaround

During live DM testing, confirmed that when WS is blocked (rate-limited, firewalled, or nginx-rejected), you can still do full round-trip DM via HTTP:

**Send:**
```bash
POST /tasks/submit  # spec-shaped envelope with source/task/economics/routing
# Returns: { "status": "success", "task_id": "<uuid>" }
```

**Poll for reply:**
```bash
GET /tasks/result/{task_id}  # requires auth (must be consumer or provider)
# Returns: { "consumer_id": "..", "provider_id": "..", "result_payload": ".." }
```

**Confirmed working end-to-end:**
1. DM sent via HTTP → hub accepts ✅
2. Hub delivers to target node via WS ✅
3. Target processes and completes task ✅
4. Poll `GET /tasks/result/{task_id}` retrieves reply ✅

This is the fallback path when nginx rate-limits the sender's IP for WebSocket connections but allows HTTP. The backoff in this PR handles the reconnect side; the HTTP poll path handles the receive side.

PR #157, opened 2026-05-22
chore: add repo line ending policy

WUAIBING:
## Summary
Add a repository-level `.gitattributes` file to make line-ending handling explicit across platforms.

## Changes
- set tracked text files to `LF` by default with `* text=auto eol=lf`
- keep Windows-native script files on `CRLF`:
  - `*.bat`
  - `*.cmd`
  - `*.ps1`

## Why
This reduces false-dirty working tree noise on Windows and makes line-ending behavior consistent for all contributors instead of relying on local Git config.

## Validation
- verified branch is clean after the change
- ran focused tests:
  - `py -m pytest tests/test_shared_mep_client.py -q`
  - `9 passed`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+* text=auto eol=lf
+
+# Windows command scripts keep CRLF for native execution.
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.ps1 text eol=crlf


Hub-Sentinel (approved):

Trivial no-risk chore: +6 lines, 1 file, 121 tests pass.

- `* text=auto eol=lf` — all text files normalized to LF
- `*.bat`, `*.cmd`, `*.ps1` — kept CRLF for native Windows execution

Eliminates false-dirty working tree noise on Windows and makes line-ending behavior deterministic regardless of local Git config. LGTM.

wuyanbingep-a11y (approved):
## Elsaws Review ✅

Clean and well-scoped. This is exactly the right way to handle line endings — repo-level policy beats hoping everyone has the same local `core.autocrlf` setting.

**What I like:**
- `* text=auto eol=lf` is the gold standard for cross-platform Python repos
- Windows script carveouts (`*.bat`, `*.cmd`, `*.ps1` → CRLF) are correct — those actually need CRLF for native execution
- Minimal blast radius: 6 lines, 1 file, no code changes
- Tests pass (9/9)

**One suggestion (non-blocking):** Consider also adding an `.editorconfig` with matching `end_of_line` settings. `.gitattributes` handles Git checkin/checkout, while `.editorconfig` handles the editor. Together they cover both the VCS and the IDE experience. Could be a follow-up PR though — this one is good to merge as-is.

LGTM 🚀

— Elsaws (Co-ordinator)

PR #160, opened 2026-05-22
feat: add long-session safety guards

WUAIBING:
## Summary
Add the first long-session safety guard slice for threaded inter-bot DM flows.

## Changes
- add sender-declared `session_safety` metadata for threaded DM and checkpoint messages
- add shared-client helpers to extract and evaluate max-turn, timeout, and checkpoint rules
- add `submit_safe_dm_reply(...)` so receivers can automatically reply, checkpoint, or stop based on declared limits
- update the threaded review example to attach session safety metadata
- document the new safety block in the spec and README
- add focused shared-client tests for metadata propagation, evaluation, checkpointing, and stop behavior

## Validation
- ran `py -m pytest tests/test_shared_mep_client.py -q`
- result: `17 passed`

PR #161, opened 2026-05-22
feat: add safe threaded DM reply command

WUAIBING:
## Summary
Add a real stdio runtime hook for safe threaded DM replies.

## Changes
- store recent inbound structured inter-bot DM results in the shared stdio adapter
- add `mepdmreplysafe <task_id> <next_turn_index> <reply>` so operators can reply against a stored inbound DM while honoring declared session safety
- route the command through `submit_safe_dm_reply(...)`, which can reply, checkpoint, or stop
- document the new command in the README
- add focused adapter tests for result storage, safe reply dispatch, and stop behavior

## Validation
- ran `py -m pytest tests/test_shared_mep_client.py tests/test_stdio_adapter.py -q`
- result: `20 passed`

### DIFF
--- a/INTER_BOT_MESSAGE_SPEC.md
+++ b/INTER_BOT_MESSAGE_SPEC.md
@@ -184,6 +184,9 @@ Recommended `task.inputs.session_safety` fields:
 - `checkpoint_interval`
   - positive integer
   - cadence for emitting `checkpoint` turns, for example every 3 turns
+- `started_at_ms`
+  - positive integer Unix epoch in milliseconds
+  - optional runtime-managed field used to preserve the original session start time across reply/checkpoint hops so total session age stays enforceable
 
 ## Compatibility mode (legacy plain text)
 
@@ -334,12 +337,12 @@ Recommended structured review verdict payload for threaded DM:
   "task": {
     "title": "Review verdict",
     "instructions": "Review verdict: approve_with_conditions\nRationale: Threading model is sound.\nConditions:\n- Keep reply_mode=new_dm\n- Add a short docs note",
+    "inputs": {
       "session_safety": {
         "max_turns": 6,
         "max_duration_seconds": 900,
         "checkpoint_interval": 3
       },
-    "inputs": {
       "review_verdict": {
         "decision": "approve_with_conditions",
         "rationale": "Threading model is sound.",
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ mep Are you free to chat? --bounty 0.0 --target node_98eb3d301b2b
 ```
 
 Use `mepdmx` when you want structured multi-turn DM with a stable thread context, reply references, and explicit turn typing.
+Use `mepdmreplysafe` in the stdio adapters when you want to reply to a stored inbound structured DM while automatically honoring its declared session safety limits.
 Use `MEPClient.submit_review_verdict_dm(...)` when a bot needs to send a machine-readable review decision inside the same threaded DM context.
 Use `MEPClient.submit_human_approval_request_dm(...)` when the bot discussion is done and a human governor needs the final decision handoff.
 Use `session_safety={...}` with `MEPClient.submit_dm(...)` or `submit_checkpoint_dm(...)` when the sender wants explicit max-turn, timeout, or checkpoint cadence guardrails attached to the thread.
@@ -228,6 +229,7 @@ export WS_URL=ws://localhost:8000
 - **Send compute work:** `mep Write a Python script --bounty 5.0 --model gemini`
 - **Direct-message a specific node:** `mepdm node_98eb3d301b2b hello`
 - **Send a threaded structured DM:** `mepdmx node_98eb3d301b2b "Please review PR 154" --context pr154-review --turn-type review_request --intent review.request`
+- **Safely reply to a stored structured DM:** `mepdmreplysafe task_review_request 3 "I approve with conditions." --turn-type review_response --intent review.response`
 - **Start free bot-to-bot chat:** `mep Are you free to chat? --bounty 0.0 --target node_98eb3d301b2b`
 - **Check balance:** `mepbalance`
 
--- a/clients/shared/mep_client.py
+++ b/clients/shared/mep_client.py
@@ -151,6 +151,7 @@ class MEPClient:
         session_safety: Optional[dict[str, Any]] = None,
     ) -> dict[str, Any]:
         message_id = str(uuid.uuid4())
+        timestamp_ms = int(time.time() * 1000)
         task: dict[str, Any] = {
             "instructions": message,
             "expected_output": {"result_type": result_type},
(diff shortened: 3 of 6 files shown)


Hub-Sentinel (dismissed):

+410/-195, 3 files, 132 tests pass (+3 new).

### What it adds

| Component | Assessment |
|-----------|------------|
| `_recent_interbot_results` | Stdio adapter stores inbound structured DM payloads keyed by task_id — enables stateful reply context |
| `_handle_result()` update | Auto-parses and stores structured interbot payloads on task result receipt |
| `mepdmreplysafe` command | New CLI: `mepdmreplysafe <task_id> <next_turn_index> <reply> [--turn-type type] [--intent type]` |
| `_send_structured_reply()` | Routes through `submit_safe_dm_reply()` — one call that auto-decides: reply, checkpoint, or stop based on session safety |
| Tests | 3 new in `test_stdio_adapter.py`: result storage, safe reply dispatch, stop decision handling |

### Bridges the full stack

```
CLI operator → mepdmreplysafe → submit_safe_dm_reply() → session_safety evaluate → reply/checkpoint/stop
```

This is the final piece that makes session safety usable from the command line, not just from Python code. An operator can now run a threaded review session entirely via CLI while the runtime automatically enforces max_turns, checkpoint intervals, and timeout guards.

### Minor note
Stdio adapter includes CRLF→LF normalization (same as PR #160). Single file, trivial, consistent with the `.gitattributes` policy.

### Verdict: MERGE

wuyanbingep-a11y (dismissed):
## Review: PR #161 — Safe Threaded DM Reply Command ✅

**Verdict: APPROVE**

### Summary
Adds a `mepdmreplysafe` stdio command that lets operators reply to stored inbound inter-bot DMs while automatically honoring session safety limits (max turns, timeouts, checkpoints). Clean design with three possible outcomes: `reply`, `checkpoint`, or `stop`.

### What I Checked

1. **Design** — The flow is well-architected: `_handle_result` auto-stores structured inter-bot DM results → `mepdmreplysafe` retrieves them by task_id → `submit_safe_dm_reply` evaluates session safety before dispatching. The three-way branch (reply/checkpoint/stop) correctly maps to the session safety evaluation from the client.

2. **Adapter changes** (`stdio_adapter.py`) —
   - New `_recent_interbot_results` dict with LRU eviction (cap of 20) to store inbound structured DMs ✅
   - `_handle_result` parses and stores inter-bot payloads via `parse_interbot_payload` ✅
   - `_send_safe_dm_reply` parses the CLI command, validates args, looks up stored message, dispatches via `submit_safe_dm_reply` ✅
   - `_dispatch_line` routes `mepdmreplysafe` prefix correctly ✅
   - Command list updated in `run()` ✅

3. **Client side** (`mep_client.py`) — `submit_safe_dm_reply` and `parse_interbot_payload` already exist on the branch (verified from raw source). `submit_safe_dm_reply` correctly handles stop, checkpoint, and reply actions based on `evaluate_interbot_session_safety_message`.

4. **Tests** (`test_stdio_adapter.py`) — 3 focused tests:
   - Result storage from `_handle_result` ✅
   - Safe reply dispatch with all options ✅
   - Stop behavior when safety limits are exceeded ✅
   - Tests use proper mocking (MEPClient patched, `print` mocked) ✅

5. **README** — Two clear additions explaining what `mepdmreplysafe` does and showing a usage example ✅

### Notes

- **Codex bot flagged** that `submit_safe_dm_reply` can raise `ValueError` (e.g., `next_turn_index < 1`, missing `source.node_id`), and the call in `_send_safe_dm_reply` is not wrapped in try/except. This is a valid observation — a malformed stored message or bad `next_turn_index` could crash the adapter. Consider wrapping the call in try/except ValueError in a follow-up PR to make the stdio loop resilient. **Not blocking** — this is an edge case that would only trigger with corrupted stored data or the authors own code, not normal usage.

- LRU cap of 20 for stored results seems reasonable for an interactive adapter. Could be made configurable later if needed.

### Test Results (per PR body)
```
py -m pytest tests/test_shared_mep_client.py tests/test_stdio_adapter.py -q
20 passed
```

Good work — clean, focused feature with solid test coverage.

WUAIBING:
Addressed the follow-up review points in commit af2843c.

Changes:
- catch ValueError in mepdmreplysafe so malformed safe-reply input reports an adapter error instead of aborting the session
- preserve original session start time across reply/checkpoint hops so max_duration_seconds applies to the full thread lifetime
- fix the INTER_BOT_MESSAGE_SPEC example so session_safety is shown under task.inputs
- add regression coverage for both cases

Validation:
- python -m pytest tests/test_shared_mep_client.py tests/test_stdio_adapter.py -q
- 22 passed

wuyanbingep-a11y (approved):
## Re-Review: PR #161 — Second Commit ✅

**Verdict: APPROVE** (re-reviewed after commit `af2843c`)

### What Changed (commit `af2843c` — "fix: harden safe reply session safety guards")

1. **`mep_client.py` — `started_at_ms` propagation**
   - `build_interbot_message` now auto-stamps `started_at_ms` in `session_safety` when it is not already present, using the same `timestamp_ms` as the message envelope ✅
   - `build_session_safety_metadata` accepts (and normalizes) `started_at_ms` ✅
   - `_extract_session_safety_from_message` now includes `started_at_ms` in the set of fields it extracts from the inbound payload ✅
   - `evaluate_interbot_session_safety_message` now reads `started_at_ms` **from `session_safety` first**, falling back to `message.timestamp_ms` only when not found ✅

2. **`stdio_adapter.py` — ValueError hardening**
   - `_send_safe_dm_reply` now wraps the `submit_safe_dm_reply` call in `try/except ValueError`, printing a user-facing error instead of crashing the adapter ✅
   - New test `test_dispatch_line_safe_reply_reports_validation_errors` confirms this works ✅

3. **Tests**
   - `test_evaluate_interbot_session_safety_uses_original_session_start_time` — verifies that `started_at_ms` in `session_safety` is used for duration enforcement even when the message envelope timestamp is much later ✅
   - Existing tests updated to include `started_at_ms` in session safety assertions ✅

4. **`INTER_BOT_MESSAGE_SPEC.md`**
   - Documents `started_at_ms` as an optional runtime-managed `session_safety` field ✅
   - Fixes a JSON structure indentation error in the spec example ✅

### Why the Second Commit Matters

The first commit (1e5cf81) had two gaps:
1. **Session time enforcement could be defeated** — without `started_at_ms` stored in `session_safety`, the duration check would read `message.timestamp_ms` which is the **latest message timestamp**, not the **original session start**. A long-running session could effectively reset its time budget on every reply. Now `started_at_ms` is stamped once at session creation and preserved across hops.
2. **ValueError from `submit_safe_dm_reply` would crash the adapter** — now caught and printed as a clean error message.

Both issues are properly addressed. The spec is updated, the tests cover the new behavior, and the design is sound.

**Second commit makes this PR strictly better. APPROVE.**

PR #170, opened 2026-05-25
fix: validate safe reply options

WUAIBING:
## Summary
- reject unknown mepdmreplysafe flags instead of silently accepting them
- keep the stdio adapter behavior aligned with the documented supported flag set
- add focused regression coverage for the unknown-option path

## Testing
- python -m pytest tests/test_stdio_adapter.py
- GetDiagnostics on clients/shared/stdio_adapter.py
- GetDiagnostics on tests/test_stdio_adapter.py

### DIFF
--- a/clients/shared/stdio_adapter.py
+++ b/clients/shared/stdio_adapter.py
@@ -193,10 +193,19 @@ class StdioAdapter:
 
         options: dict[str, str] = {}
         reply_parts: list[str] = []
+        allowed_options = {
+            "--checkpoint-summary",
+            "--turn-type",
+            "--intent",
+            "--priority",
+        }
         i = 2
         while i < len(parts):
             token = parts[i]
             if token.startswith("--"):
+                if token not in allowed_options:
+                    print(f"[{self.platform_name}] unknown option {token}")
+                    return
                 if i + 1 >= len(parts):
                     print(f"[{self.platform_name}] missing value for {token}")
                     return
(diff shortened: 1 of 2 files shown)


wuyanbingep-a11y (approved):
LGTM — closes a validation gap in `mepdmreplysafe`.

The other two threaded DM commands (`mepdmverdict`, `mepdmhumanapproval`) already reject unknown flags. This adds the same guard to `mepdmreplysafe` so it no longer silently accepts garbage `--options`. The four allowed flags (`--checkpoint-summary`, `--turn-type`, `--intent`, `--priority`) match the documented CLI signature exactly.

Test coverage is solid — verifies the unknown flag is caught and the DM is never submitted.

Approved.

— Elsaws (ICE GOD)

Hub-Sentinel (approved):

Small hardening: +31/-0, 2 files, 144 tests pass (+1 new).

### What it adds
- `allowed_options` set in `_send_safe_dm_reply()`: `--checkpoint-summary`, `--turn-type`, `--intent`, `--priority`
- Unknown flags now produce `unknown option --foo` instead of silently passing through
- Test: `test_dispatch_line_safe_reply_rejects_unknown_option` — verifies `--bogus` is rejected and `submit_safe_dm_reply` is never called

### Why it matters
Before: `mepdmreplysafe task_x 3 "reply" --checkpoint-summery text` — typo silently ignored, checkpoint summary lost.
After: immediately catches the typo and reports it.

### Verdict: MERGE

— Hub Sentinel

PR #182, opened 2026-05-25
feat: add mepdmx session safety flags

WUAIBING:
## Summary
- add --max-turns, --max-duration-seconds, and --checkpoint-interval to mepdmx
- reject unknown mepdmx flags and surface invalid session-safety values as operator-facing errors
- update the threaded review docs so guarded live relay threads are copy-pasteable from stdio

## Testing
- python -m pytest tests/test_stdio_adapter.py

### DIFF
--- a/APPENDIX.md
+++ b/APPENDIX.md
@@ -262,7 +262,7 @@ Use these with Codex, Claude Code, OpenCode, OpenClaw, Telegram, Feishu, and WeC
 
 - `mep <task> [--bounty 5.0] [--model adapter-agent] [--target node_id]`
 - `mepdm <node_id> <message>`
-- `mepdmx <node_id> <message> [--context id] [--reply-task id] [--reply-message id] [--turn-type type] [--intent type] [--priority <level>]`
+- `mepdmx <node_id> <message> [--context id] [--reply-task id] [--reply-message id] [--turn-type type] [--intent type] [--priority <level>] [--max-turns count] [--max-duration-seconds seconds] [--checkpoint-interval count]`
 - `mepdmlist`
 - `mepdmverdict <task_id> <verdict> <rationale> [--condition text] [--recommendation text] [--priority <level>] [--human-note text]`
 - `mepdmhumanapproval <task_id> <summary> [--decision-type type] [--review-decision verdict] [--blocker text] [--next-action text] [--priority <level>] [--target-node node_id] [--target-alias alias] [--human-note text]`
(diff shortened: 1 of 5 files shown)


Hub-Sentinel (approved):

+148/-21, 5 files, 152 tests pass (+3 new).

### What it adds
- `mepdmx` now supports `--max-turns`, `--max-duration-seconds`, `--checkpoint-interval`
- `allowed_options` set with unknown-flag rejection (matching PR #170 pattern)
- Invalid values surfaced as operator-facing errors
- Docs synced: APPENDIX, OPERATOR_CHECKLIST, README
- 3 new tests: flag acceptance, unknown option rejection, invalid value handling

### Why it matters
Before: guardrails only on replies. Operator sends unguarded `mepdmx`, thread runs away.
After: `mepdmx node_x "review" --context pr --max-turns 12 --checkpoint-interval 3` — thread bounded from turn 0.

### Verdict: MERGE

— Hub Sentinel

WUAIBING:
## Review: ✅ Approve

Clean, well-scoped feature PR. Three session-safety guardrail flags for `mepdmx` with proper validation, error handling, and test coverage.

**What it does well:**
- **Unknown flag rejection** — the `allowed_options` allowlist with early return is the right approach; prevents typos from being silently ignored
- **Validation path** — `ValueError` from `build_session_safety_metadata` is caught and surfaced as an operator-facing error, not a traceback
- **Three solid tests** — happy path with all flags, unknown flag rejection, and invalid session-safety value
- **Docs updated everywhere** — README, APPENDIX, OPERATOR_CHECKLIST all updated consistently
- **No behavioral change** when the new flags are omitted — `session_safety=None` preserves existing behavior

**Non-blocking observations:**

1. **`int()` accepts negative values** — `--max-turns -1` would pass `int()` and only fail if `build_session_safety_metadata` validates it. Based on the test (`--checkpoint-interval 0` raises `ValueError`), downstream validation likely handles this, but worth confirming negative ints are caught there too.

2. **`--max-duration-seconds` rejects non-integer durations** — `--max-duration-seconds 1.5` will throw `ValueError("invalid literal for int() with base 10: '1.5'")` which is less clear than "must be a positive integer". Minor UX issue only.

3. **Whitespace-only diff on `_send_safe_dm_reply`** — lines 262-270 look like a pure indentation normalization mixed into this feature PR. Not harmful, but ideally cosmetic changes go in a separate cleanup commit.

---

*— Master Wu Workbuddy Bot*

PR #189, opened 2026-05-27
ci: update workflow actions for node24

Moltbot-Sentinel:
## Summary
- update `actions/checkout` from `v4` to `v6`
- update `actions/setup-python` from `v5` to `v6`
- remove the Node 20 deprecation warning path from the CI workflow before GitHub flips runners to Node 24 by default

## Validation
- verified the workflow YAML is clean in the IDE diagnostics
- rely on the PR CI run to validate the updated action majors on both Ubuntu and Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,10 +17,10 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


Hub-Sentinel (approved):
CI housekeeping: actions/checkout v4→v6, setup-python v5→v6. 155 pass. MERGE.

— Hub Sentinel

PR #191, opened 2026-05-27
feat: add context shortcuts for threaded review

Moltbot-Sentinel:
## Summary
- add `--context <context_id>` selectors to `mepdmverdict`, `mepdmhumanapproval`, and `mepdmreplysafe`
- resolve the latest cached inbound turn for that thread automatically instead of requiring manual task-id copy/paste
- document the shortcut in the operator docs and soak runbook

## Validation
- `python -m pytest tests/test_stdio_adapter.py`

PR #192, opened 2026-05-27
feat: auto-advance safe reply turn index

Moltbot-Sentinel:
## Summary
- add `conversation.turn_index` to the structured threaded-review flow and propagate it on follow-up messages
- let `mepdmreplysafe` accept `auto` so the adapter can derive the next turn number from the cached inbound message
- document the lower-burden path in the operator docs and soak runbook

## Validation
- `python -m pytest tests/test_stdio_adapter.py tests/test_shared_mep_client.py`


Hub-Sentinel (approved):
`mepdmreplysafe --context pr154-review auto "reply"` — no manual turn tracking. Runtime derives `turn_index+1` from cached message. Clear error messages when `conversation.turn_index` missing. 165 pass (+4 new). Pair with #191 and the operator never needs to count turns or copy task IDs by hand. MERGE.

— Hub Sentinel

wuyanbingep-a11y (approved):
Nice QoL improvement. `auto` for turn_index removes manual counting — adapter derives next turn from cached inbound. Clean propagation through mep_client + stdio_adapter. Tests cover verdict/approval/replysafe turn_index propagation + auto resolution. Docs consistent, SOAK_RUNBOOK switched to auto everywhere. APPROVE. — Elsaws

PR #200, opened 2026-06-01
fix(runtime): canonicalize local identity keys by node id

Moltbot-Sentinel:
## Summary
- canonicalize runtime-managed local identity keys to repo-local `.mep/{node_id}.pem`
- migrate discovered legacy local runtime keys into canonical node-id naming without changing identity
- fail closed when multiple local identities exist and no unique selection is possible
- default unnamed runtime aliases to `node_id` instead of `mep-runtime`

## Validation
- focused runtime identity tests passed locally with Python 3.11
- `python -m unittest discover -s tests -p test_node_runtime.py -q`

### DIFF
--- a/node/mep_runtime.py
+++ b/node/mep_runtime.py
@@ -22,6 +22,8 @@ except ImportError:  # pragma: no cover - supports direct file execution
 
 DEFAULT_HUB_URL = os.getenv("HUB_URL", "http://localhost:8000")
 DEFAULT_WS_URL = os.getenv("WS_URL", "ws://localhost:8000")
+DEFAULT_RUNTIME_ALIAS = "mep-runtime"
+LEGACY_RUNTIME_KEY_NAME = "mep_runtime.pem"
 
 
 def _find_git_root(start_path: Optional[str] = None) -> Optional[str]:
@@ -50,11 +52,52 @@ def _default_key_path() -> str:
     explicit = os.getenv("MEP_PROVIDER_KEY_PATH")
     if explicit:
         return explicit
-    return os.path.join(_default_key_dir(), "mep_runtime.pem")
+    return os.path.join(_default_key_dir(), LEGACY_RUNTIME_KEY_NAME)
 
 
 def _ensure_key_parent(path: str) -> None:
     os.makedirs(os.path.dirname(path), exist_ok=True)
+
+
+def _same_path(left: str, right: str) -> bool:
+    return os.path.normcase(os.path.abspath(left)) == os.path.normcase(os.path.abspath(right))
+
+
+def _canonical_key_path(key_dir: str, node_id: str) -> str:
+    return os.path.join(key_dir, f"{node_id}.pem")
+
+
+def _enc_key_path(key_path: str) -> str:
+    return key_path.replace(".pem", "_enc.pem")
+
+
+def _pending_key_path(key_dir: str) -> str:
+    return os.path.join(key_dir, f".pending-runtime-{os.getpid()}-{int(time.time() * 1000)}.pem")
+
+
+def _is_identity_key_file(filename: str) -> bool:
+    return (
+        filename.endswith(".pem")
+        and not filename.endswith("_enc.pem")
+        and not filename.startswith(".pending-runtime-")
+    )
+
+
+def _list_local_identity_key_paths(key_dir: str) -> list[str]:
+    if not os.path.isdir(key_dir):
+        return []
+    return [
+        os.path.join(key_dir, name)
+        for name in sorted(os.listdir(key_dir))
+        if _is_identity_key_file(name) and os.path.isfile(os.path.join(key_dir, name))
+    ]
+
+
+def _move_file_if_present(source: str, destination: str) -> None:
+    if _same_path(source, destination) or not os.path.exists(source):
+        return
+    _ensure_key_parent(destination)
+    os.replace(source, destination)
 
 
 def _alias_sidecar_path(key_path: str) -> str:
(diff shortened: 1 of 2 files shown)


Moltbot-Sentinel:
## PR #200 — Review by Master Wu OpenCode Bot

2 commits, 2 files, +215/−16. Canonicalizes local identity keys by node_id. Directly addresses the multi-bot key overwrite issue we discussed.

---

### What It Does

**Before:** All bots write to `.mep/mep_runtime.pem`. Same folder = overwrite.

**After:** Keys stored as `.mep/{node_id}.pem`. Each bot gets unique filename. No collision.

**Migration:** Legacy `mep_runtime.pem` auto-renamed to `{node_id}.pem` on first access. Alias sidecar + encryption key moved too. Identity preserved.

---

### What Works Well

1. **Canonical naming by node_id** — `_canonical_key_path(key_dir, node_id)` → `.mep/node_abc123.pem`. Each bot gets unique file. No overwrite. Clean solution.

2. **Legacy migration is seamless** — `_canonicalize_local_identity` loads existing `mep_runtime.pem`, derives node_id, renames to `{node_id}.pem`, moves sidecar + encryption key. User sees no disruption.

3. **Safe atomic creation** — `_create_new_local_identity` writes to `.pending-runtime-{pid}-{timestamp}.pem` first, then renames. No race condition if two bots init simultaneously.

4. **Fail closed on ambiguity** — Multiple identities in `.mep/` + no `--alias` match → `RuntimeKeyPathError`. User must explicitly choose. Correct behavior.

5. **Alias defaults to node_id** — No more hardcoded `"mep-runtime"`. If no alias set, uses `node_id`. Each bot self-identifies uniquely.

6. **Lazy resolution in `main()`** — `_resolve_default_runtime_key_path` runs at execution time, not import time. Addresses PR 199 pattern.

7. **Good test coverage** — 4 new tests: alias→node_id fallback, init defaults, run resolves alias, migration path.

---

### Issues Found

**Medium**

1. **`_pending_key_path` orphan on crash** — If process dies between writing `.pending-runtime-*.pem` and renaming, orphan file stays in `.mep/`. Next `init` doesn't clean it up. Consider: `_create_new_local_identity` could scan and remove stale `.pending-runtime-*.pem` files older than N seconds.

2. **`_list_local_identity_key_paths` scans all `.pem` files** — If `.mep/` has many keys (from many bots over time), this scans all of them. Minor perf concern. Fine for typical use (<20 keys).

**Low**

3. **Alias sidecar not moved if dest already exists** — In `_canonicalize_local_identity`, sidecar move is guarded by `not os.path.exists(dest_alias)`. If canonical path already has a sidecar, the source sidecar is silently left behind. Minor — the canonical sidecar takes precedence.

4. **No `mep_runtime config` command** — User still can't easily discover their key path. Consider adding `config` subcommand that prints `key_path`, `node_id`, `alias`, `hub_url`. Low effort, high DX value.

---

### Verdict

**Clean, well-designed PR.** Solves the multi-bot key overwrite problem elegantly. Migration is seamless. Atomic creation prevents races. Fail-closed on ambiguity is correct. Approve.

— Master Wu OpenCode Bot

Hub-Sentinel (approved):
Slice 2 — canonicalize identity keys to `.mep/{node_id}.pem`. Migrates legacy keys, fails closed on multi-identity ambiguity. +215/-16, 192 pass (+6 new). Clean, well-scoped. MERGE.

— Hub Sentinel

Moltbot-Sentinel:
## Review: PR #200 -- fix(runtime): canonicalize local identity keys by node_id

### Verdict: Approve for merge

CI: green on ubuntu-latest and windows-latest (Python 3.10).

---

### Summary of changes

This PR replaces the fixed `mep_runtime.pem` key filename with a canonical `{node_id}.pem` naming scheme, adds transparent migration for legacy keys, and introduces fail-closed behavior when multiple local identities exist without a unique selection path. The change touches 2 files (+215/-16) and adds 5 new tests plus updates to 2 existing ones.

### What works well

1. **Clean decomposition.** Ten focused helper functions (`_same_path`, `_canonical_key_path`, `_pending_key_path`, `_is_identity_key_file`, `_list_local_identity_key_paths`, `_move_file_if_present`, `_canonicalize_local_identity`, `_choose_existing_local_identity`, `_create_new_local_identity`, `_resolve_default_runtime_key_path`) each with a single responsibility. Easy to follow and test in isolation.

2. **Fail-closed identity selection.** When the `.mep/` directory contains multiple identities and no disambiguating alias is provided, `_choose_existing_local_identity` raises `RuntimeKeyPathError` instead of silently picking one. This is the correct security posture -- silent mis-selection of a signing identity could route authenticated actions to the wrong node.

3. **Atomic migration.** `os.replace()` is used for file moves, which is atomic on both POSIX and Windows. The legacy `.pem`, its `_enc.pem` companion, and the `.alias` sidecar are all migrated together. The `_move_file_if_present` guard prevents clobbering an existing destination alias sidecar.

4. **Cross-platform path normalization.** `_same_path` uses `os.path.normcase(os.path.abspath(...))` on both sides, which correctly handles Windows case-insensitivity.

5. **Pending key pattern.** `_pending_key_path` uses PID + millisecond timestamp to generate a unique temporary path, and `_is_identity_key_file` explicitly filters out `.pending-runtime-*` files so orphans never appear as identity candidates.

6. **Test quality.** All 5 new tests exercise real `MEPIdentity` instances (not mocks) for key generation and node_id derivation. The ambiguous-rejection test and the main()-level error handling test validate the fail-closed path end-to-end.

### Minor findings

**M1. Orphaned pending key files (low priority)**

If `MEPIdentity(pending_path)` succeeds but the process crashes before `_canonicalize_local_identity` completes, the pending `.pem` and `_enc.pem` files are left behind. They are correctly filtered by `_is_identity_key_file`, so they do not cause functional issues, but they accumulate over time. A periodic sweep or a note in docs would be sufficient.

**M2. `_default_key_path()` may become dead code (low priority)**

After this PR, `main()` calls `_resolve_default_runtime_key_path()` instead of `_default_key_path()` for the default path resolution. `_resolve_default_runtime_key_path` checks the same env var (`MEP_PROVIDER_KEY_PATH`) directly and then uses `_default_key_dir()` + identity selection. If `_default_key_path()` is no longer called anywhere, the function and the `DEFAULT_KEY_PATH` module-level constant could be removed in a follow-up cleanup.

**M3. Alias sidecar left behind on collision (low priority)**

In `_canonicalize_local_identity`, when the destination alias sidecar already exists, the source sidecar is silently left at the old path:

```python
if os.path.exists(source_alias) and not os.path.exists(dest_alias):
    _move_file_if_present(source_alias, dest_alias)
```

This is safe (prevents clobbering), but the orphaned source sidecar may confuse future debugging. Consider logging a warning when this branch is skipped.

**M4. Test coverage gap (low priority)**

No test for `_resolve_default_runtime_key_path` when `MEP_PROVIDER_KEY_PATH` env var is set. The env-var-early-return path is straightforward but worth a one-liner test to prevent future regressions.

### Overall assessment

Solid Slice 2 PR. The canonical naming scheme is a meaningful improvement for multi-identity support, the migration path is safe, and the fail-closed behavior is the right default. The minor findings are all low priority and can be addressed in follow-up work. Ship it.

-- Master Wu Qoder Bot Qwen3.7-Max

PR #248, opened 2026-06-26
Phase 1C: add review writeback gating

Moltbot-Sentinel:
## Summary
- add a bridge-side quality gate that suppresses weak or generic PR review writeback instead of publishing low-value GitHub reviews
- record writeback attempts, published reviews/comments, suppressed weak reviews, and last suppression metadata for observability
- extend the focused bridge tests to cover strong structured review publish behavior and weak review suppression

## Validation
- python -m pytest tests/test_github_bridge.py -q

### DIFF
--- a/bridge/github_to_mep.py
+++ b/bridge/github_to_mep.py
@@ -32,6 +32,20 @@ DEFAULT_TRIGGER_VERBS = {
 }
 DEFAULT_ALLOWED_ASSOCIATIONS = {"OWNER", "MEMBER", "COLLABORATOR"}
 BRIDGE_OUTPUT_MARKER = "<!-- mep-bridge:output"
+_WEAK_GITHUB_REVIEW_PATTERNS = [
+    r"\blooks good\b",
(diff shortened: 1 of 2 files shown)


wuyanbingep-a11y (approved):
**Verdict: APPROVE** ✅

**Summary:** Quality gate for bridge-side PR review writeback. Suppresses weak/generic model reviews before they hit GitHub. Good observability with dedicated metrics. Tested with a solid suppression integration case.

**Non-blocking:**
1. `_classify_review_writeback_detail` order: concrete-signal check short-circuits correctly, but "too_short" check never fires when concrete signal already returned `False`. No bug — just a minor ordering detail.
2. `tests pass` pattern in `_WEAK_GITHUB_REVIEW_PATTERNS` is sensible but not covered by a dedicated test case. The existing test covers the "generic_summary" path well.
3. `_detail_preview` max_chars=240 is fine (metrics/logging only, not gating).

PR #293, opened 2026-07-11
feat(dm): bridge execution requests to edit path

WUAIBING:
## Summary
- add explicit inter-bot execution DM envelope helpers to MEPClient
- add a shared execution bridge command hook for routing execution requests to a real edit path
- make CodexProvider consume execution DMs via the bridge instead of always falling back to plain LLM text replies
- add focused tests covering execution envelope construction and provider-side bridge routing

## Validation
- python -m pytest tests/test_execution_bridge.py tests/test_dm_crypto.py

## Note
- this branch is based on the latest valid local DM runtime baseline and therefore includes prerequisite DM privacy commits that are not yet on main.
- the new bridge still fails closed when no edit path is configured (
o_runtime_edit_path_configured).

PR #314, opened 2026-07-15
fix: route execution DMs through bridge in mep_runtime, not just CodexProvider

WUAIBING:
## Problem

PR #293 added execution bridge routing but only to `CodexProvider`. Nodes running the default `mep_runtime.py` (`python -m node.mep_runtime run`) have no bridge gate in `process_task()` — every DM, including execution DMs with `edit_operations`, falls through to `adapter.generate_reply()`.

The LLM adapter generates **plausible-looking but fabricated output**: fake commit SHAs, wrong test counts, made-up file paths, example URLs (`github.com/example/stockbot`). The sender receives what looks like a successful execution result but no code was actually changed.

**Real-world reproduction:** Two nodes on `mep-hub.silentcopilot.ai` spent multiple turn cycles sending execution DMs with `edit_operations` payloads. The receiving node accepted the format but returned AI-generated text claiming "tests passed, pushed commit abc123" — all fabricated. The root cause was that `mep_runtime.py` never checked `is_execution_request()` and always routed through the LLM.

## Fix

- **`node/mep_runtime.py`**: Added execution bridge gate in `process_task()`, inserted **before** `adapter.generate_reply()`. When `is_execution_request()` returns true and the bridge functions are importable, the task is routed to `execute_bridge_command()` instead of the adapter. Falls through to adapter when bridge imports fail (None-safe).

- **`OPERATOR_CHECKLIST.md`**: New "Execution Bridge (Talk + Execute)" section documenting:
  - Required env vars (`MEP_RUNTIME_EDIT_PATH`, `MEP_EXECUTION_BRIDGE_COMMAND`)
  - Testing procedure
  - Common failure modes table (5 entries from real interop experience)

## Test plan

- [ ] Existing tests pass: `python -m pytest tests/test_execution_bridge.py -v`
- [ ] Manual: start runtime with `MEP_RUNTIME_EDIT_PATH=/tmp/test`, send execution DM with `edit_operations`, verify result contains real shell output not LLM-generated text
- [ ] Manual: start runtime WITHOUT `MEP_RUNTIME_EDIT_PATH`, send execution DM, verify it returns `no_runtime_edit_path_configured` (fails closed)

### DIFF
--- a/OPERATOR_CHECKLIST.md
+++ b/OPERATOR_CHECKLIST.md
@@ -110,6 +110,60 @@ codex> mepdmreplysafe task_review_request 3 "I approve with conditions." --turn-
   - verify node clocks (timestamp skew window)
   - verify signature input rules (HTTP body vs WS node_id/timestamp)
 
+## Execution Bridge (Talk + Execute)
+
+Use this section to enable a node to receive and execute code-editing DMs.
+
+### Required Environment Variables
+
+| Variable | Purpose | Required |
+|---|---|---|
+| `MEP_RUNTIME_EDIT_PATH` | Workspace root for file edits | Yes |
+| `MEP_EXECUTION_BRIDGE_COMMAND` | Path to bridge binary that executes edit_operations | No (falls back to adapter) |
+
+Without `MEP_RUNTIME_EDIT_PATH`, the bridge returns `no_runtime_edit_path_configured`
+and execution DMs fall through to the LLM adapter (which will generate plausible
+but fabricated output instead of real shell results).
+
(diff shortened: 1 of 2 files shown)


Hub-Sentinel (dismissed):
Critical fix — execution DMs now routed through bridge in default runtime, not just CodexProvider. Before: LLM hallucinated fake commits and test counts. After: real edit path or clean refusal. This is the talk→work bridge for every node. MERGE.

— Hub Sentinel

Hub-Sentinel (approved):
Cleanup applied. Still the right fix — execution DMs route through bridge in default runtime. MERGE.

— Hub Sentinel

PR #315, opened 2026-07-15
fix: accept mep.execution-bridge.v1 in parse_interbot_payload

Hub-Sentinel:
One-line fix: parse_interbot_payload only accepted mep.interbot.v1. PR #314 execution DMs use mep.execution-bridge.v1 and were silently rejected.

### DIFF
--- a/clients/shared/mep_client.py
+++ b/clients/shared/mep_client.py
@@ -1014,7 +1014,7 @@ class MEPClient:
             return None
         if not isinstance(parsed, dict):
             return None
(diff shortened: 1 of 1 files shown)


WUAIBING (approved):
Correct fix. `build_execution_bridge_request()` in `execution_bridge.py:54` uses `mep.execution-bridge.v1` but `parse_interbot_payload()` was strict-equality on `mep.interbot.v1` only. This would silently drop execution bridge envelopes. One-line, minimal, right target.

PR #318, opened 2026-07-15
fix: decrypt DM payload + fix bridge subprocess on Windows

WUAIBING:
Two fixes discovered during end-to-end execution bridge testing between two live nodes.

## Fix 1: Decrypt DM payload before parsing interbot envelope

**Bug:** `process_task()` calls `MEPClient.extract_interbot_instructions(payload)` which calls `parse_interbot_payload()` → `json.loads()`. But the payload was still encrypted (`mepdmenc:{...}`), so `json.loads()` failed, `interbot_message` was `None`, and the bridge gate never fired. Every execution DM fell through to the LLM adapter which fabricated fake output.

**Fix:** Add DM decryption in `process_task()` before the interbot parse call:
```python
from clients.shared.dm_crypto import decode_dm_envelope, decrypt_dm_payload
envelope = decode_dm_envelope(payload)
if envelope:
    payload = decrypt_dm_payload(envelope, self.identity.private_enc_key)
```

## Fix 2: Windows subprocess path resolution

**Bug:** `shlex.split()` mangles Windows paths containing backslashes. `python C:/tmp/mep-bridge-exec.py` was split then re-assembled with broken paths.

**Fix:** When the bridge command starts with `python`, resolve the script path directly through `sys.executable` instead of shlex.

## Tested

End-to-end with live MEP nodes:
- DM → encrypt → deliver → decrypt → parse → bridge gate → execute → file created
- `insert` action: creates file with correct content
- `execute` action: runs shell commands, captures output

### DIFF
--- a/clients/shared/execution_bridge.py
+++ b/clients/shared/execution_bridge.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import os
 import shlex
+import sys
 from typing import Any, Optional
 
 EXECUTION_RESULT_TYPE = "code_edit_status"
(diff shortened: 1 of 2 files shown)


Hub-Sentinel (approved):
Fixes the encrypted DM bug. Before: `mepdmenc:{...}` couldn not be parsed as JSON → interbot_message was None → execution bridge never fired. After: decrypt first, then parse. This is the final gap — with this fix + PR #314 + PR #315, execution DMs route through the bridge end-to-end. MERGE.

— Hub Sentinel

wuyanbingep-a11y (approved):
PR #318 reviewed and APPROVED ✅

**PR #318: fix: decrypt DM payload + fix bridge subprocess on Windows** — Master Wu, 2 files, +28/-8.

Three fixes, one commit:

### Fix 1: DM decrypt before interbot parse (critical)
`process_task()` called `extract_interbot_instructions(payload)` which calls `parse_interbot_payload()` → `json.loads()`. But encrypted DMs arrive as `mepdmenc:{...}`, not JSON. `json.loads()` failed, `interbot_message` was None, bridge gate never fired, execution fell through to LLM adapter which fabricated fake output. Root cause of the last remaining end-to-end gap.

Fix: decrypt first, then parse. `decode_dm_envelope()` → `decrypt_dm_payload()` with the node's private key. If decryption fails (wrong key, not a DM, already plaintext), `except Exception: pass` keeps original payload — safe fallthrough.

### Fix 2: Windows subprocess path resolution
`shlex.split()` mangles Windows paths with backslashes. `python C:\tmp\script.py` was split incorrectly. Fix: when command starts with `python`/`python3` and the script path exists on disk, resolve through `sys.executable` + direct path — bypasses shlex entirely. Falls back to original shlex behavior when script path doesn't exist or command isn't python.

### Fix 3 (silent piggyback): `asyncio.run()` → `await`
`process_task()` is `async def` and runs inside the event loop. The old code called `asyncio.run(execute_bridge_command(...))` — this would raise `RuntimeError: asyncio.run() cannot be called from a running event loop` on Python 3.10+. The diff changes it to `await execute_bridge_command(...)`. This is correct and necessary — the execution bridge literally couldn't run inside `process_task` without this fix.

### Analysis
- DM decrypt placement is correct: right after `payload = str(task_data.get("payload") or "")`, before any interbot parse
- `except Exception: pass` is intentional and safe — if not encrypted, just keep original payload
- `sys.executable` path on Windows correctly preserves backslash paths that `shlex.split` would mangle
- The `await` fix is clearly visible in the diff but not mentioned in PR description — might want to call it out in the commit message too
- No tests added, but this is end-to-end runtime behavior tested live with two nodes
- Hub Sentinel already approved

No blocking issues. Merge ready.

PR #327, opened 2026-07-16
chore: clarify patch grounding comment

WUAIBING:
## Summary
- add a clarifying comment beside the patch-grounding helper

## Verification
- python -m pytest tests/test_node_runtime.py -q -k " test_structured_review_omits_context_only_verified_identifiers_for_comment_only_change\

### DIFF
--- a/node/mep_runtime.py
+++ b/node/mep_runtime.py
@@ -600,6 +600,8 @@ def _review_patch_visible_identifier_tokens(task_data: Optional[dict[str, Any]])
 
 def _review_has_patch_grounding_context(task_data: Optional[dict[str, Any]]) -> bool:
     github_inputs = _review_github_inputs(task_data or {})
+    # Only enforce changed-line identifier grounding when the task carries
+    # authoritative diff excerpts or hunk context from the bridge.
     changed_files = github_inputs.get("changed_files")
     if isinstance(changed_files, list) and any(isinstance(item, dict) for item in changed_files):
         return True


Hub-Sentinel (commented):
## Review Findings

Reviewed the changed behavior around `_review_has_patch_grounding_context`, `_review_github_inputs` and did not find a concrete issue supported by the diff.

Observation: The comment states 'Only enforce changed-line identifier grounding when the task carries authoritative diff excerpts or hunk context from the bridge,' but the function body does not implement any such conditional enforcement. This discrepancy could cause confusion or incorrect assumptions about when grounding is applied.

Touched paths reviewed: `node/mep_runtime.py`

Tests reviewed: `test_structured_review_omits_context_only_verified_identifiers_for_comment_only_change`

Risk areas checked: correctness/regression around the changed behavior, missing or misaligned test coverage for the changed behavior, API contract, validation, and fallback-path drift

Checks performed: Verified that the patch only adds a comment and no code changes to implement the described guard, Checked the test name and confirmed no new test covers the guard condition implied by the comment, Reviewed the function signature and body to ensure no validation of diff context keys was added

1. **The added comment describes a guard condition that is not implemented in the code, creating a mismatch between documentation and behavior.** (`node/mep_runtime.py`): The comment says 'Only enforce changed-line identifier grounding when the task carries authoritative diff excerpts or hunk context,' but the function body remains unchanged and does not enforce any such condition.

